PR #10, opened 2014-08-17
[RFC] Singularity

igorw:
This pull request serves as a place for discussing some radical changes to the react API that I've been playing with. This is just a concept. It's just an idea. It needs a lot more work before the rest of react could use it.

Feel free to share your thoughts here though.


src/Timer/Timer.php
```diff
@@ -59,11 +52,11 @@ public function isPeriodic()
 
     public function isActive()
     {
-        return $this->loop->isTimerActive($this);
+        return l\loop()->isTimerActive($this);
```
clue: IMHO it makes sense to no longer expose the loop as part of the timer interface (i.e. drop the `getLoop()` method).

Though arguably, I'm also pro exposing the one loop via global state.

However, I consider this very change a design flaw. Unless we move this logic to the timer itself, it *actually does* have a dependency on the loop. Accessing global state here means we're using a leaky abstraction and hiding the dependencies this particular class has. Also, this does make testing harder (no built-in mechanism to mock the loop).

clue:
I agree, this appears to one of the few places where it actually makes sense to add global state. If only for convenience (and hence a cleaner design for components building on top of this) :+1: 

I've toyed with the concept of combing several loops for some time already, but even then there's only ever a single "top" loop.

So while I agree it makes sense to add a global accessor (let's face it, we're talking about a beloved singleton), I tend to disagree with actually *using* it in any of our methods (except for constructors).

mbonneau:
:+1:  - Would also like to see the loop start automatically.

src/Factory.php
```diff
@@ -1,21 +0,0 @@
-<?php
-
-namespace React\EventLoop;
-
-class Factory
-{
-    public static function create()
```
mbonneau: Could this stay and just ```return loop();``` ?

mbonneau:
This PR has been here for well over a year. Anyone who would like to comment has had the opportunity. Can we get a yes or no on this?

mdrost:
Is there any case when having two loops is helpful?

PR #28, opened 2015-04-21
clean up some duplicated code related to TickQueue

drealecs:


cboden:
`nextTick` and `futureTick` have different functionality.

drealecs:
Yes, the objects have different purpose. However, the implementation is the same...

daltones:
Have a look at method `tick`. Different functionality AND implementation. Maybe you didn't understand the difference.

drealecs:
Indeed, now I see it. Sorry for your time. I'll just close this now...

drealecs:
The documentation might not have been to clear or maybe it was just me.
If we're here, we might as well ask try to figure it out why I was confused in the implementation interface (excepting the fact that I didn't pay attention when I should have).

Is there any contract regarding calling `nextTick` from a callback that was scheduled with a `futureTick `? Because callbacks registered this way will be executed after the timers and this is not expected as declared in the documentation.

Is there any contract regarding calling `futureTick` from a callback that was scheduled with a `nextTick`? Because callbacks registered this way will be executed before any timers and this might not be expected.

Thank you!

PR #38, opened 2015-11-14
change properties's access privilege to protected.

huyanping:
in order to get read stream, write stream and timer count through inheritance. like:$loop->getReadStreamCount();

### DIFF
--- a/src/ExtEventLoop.php
+++ b/src/ExtEventLoop.php
@@ -16,17 +16,17 @@ use SplObjectStorage;
  */
 class ExtEventLoop implements LoopInterface
 {
-    private $eventBase;
-    private $nextTickQueue;
-    private $futureTickQueue;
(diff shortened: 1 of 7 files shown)


cboden:
What do you need access to these for?

cebe:
> in order to get read stream, write stream and timer count through inheritance. like:$loop->getReadStreamCount();

If you have the need to such a method it would be better to send a PR adding it instead of making all properties protected...

clue:
> in order to get read stream, write stream and timer count through inheritance. like:$loop->getReadStreamCount();

You can also build a decorating class that merely keeps track of these via composition.

Perhaps https://github.com/WyriHaximus/reactphp-event-loop-inspector is also a good start?

huyanping:
@clue,
Thx, that is a good way to solve the problem.

Issue #55, opened 2016-08-19
Recommanded Loop implementation for php7


tkleinhakisa:
Hi,

i'm trying to upgrade my app to php7. I used libevent and the matching pecl extension which is not available anymore for php7

What would you recommend ?
- use the libevent extension fork from https://github.com/expressif/pecl-event-libevent ?
- use LibEvLoop (https://github.com/m4rw3r/php-libev) last commit is from 2013 so i doubt it is ready for php7
- use the pecl event extension which seems to be ready for php7 (i have not tried) https://pecl.php.net/package-changelog.php?package=event&release=2.1.0

Did some of you make the switch already ?
What solution did you choose ?

Thank you for the effort in ReactPHP

geki-yaba:
I used expressif's version with a patch to make pecl-libevent work with php7. Without the patch my server instance would just crash on startup. Though, causing one memory leak.

Today I switched to pecl-event-2.2.1 (master snapshot) for its issue 31 [0]. pecl-event supports php7 since version 2.1.0. For now, it looks promising; the server instance started and could be terminated with ctrl-c. So, running as a daemon works, which only works with a libevent backend.

[0] https://bitbucket.org/osmanov/pecl-event/issues/31

leric:
There's another libev binding for php in pecl which is actively maintained, and php7 compatible, could be a good choice for another event-loop implementation.
http://pecl.php.net/package/ev

clue:
The libevent extension does not currently support PHP7. There's ongoing effort to update libevent to support PHP7 in php/pecl-event-libevent#2. However, the current version occasionally segfaults. This could not be observed in earlier version, so it is believed that this is a bug in libevent, rather than React.
As such, there's little we can do here. Discussion about this is probably best directed at https://github.com/reactphp/event-loop/issues/40.

In the meantime, you may want to check out the "event" extension. We've seen reports of people successfully using this on PHP7.

clue:
> In the meantime, you may want to check out the "event" extension. We've seen reports of people successfully using this on PHP7.

I'll assume this is resolved and will close this for now, please feel free to report back otherwise :+1:

rhclayto:
I created a pull request adding support for pecl-ev (http://pecl.php.net/package/ev): 

https://github.com/reactphp/event-loop/pull/83

Issue #40, opened 2015-12-24
[libevent] is currently not work in php7


kocoten1992:
It all start with:
### /vendor/react/event-loop/LibEventLoop.php on line 34

After trying to make it work, it come to:
### pecl/libevent requires PHP (version >= 5.3.0, version <= 6.0.0, excluded versions: 6.0.0), installed version is 7.0.1-2+deb.sury.org~wily+1

No valid packages found

Not sure if I report this here is right since it depend on libevent

majuca:
Hi,

I can run with php 7 by compiling the libevent : https://github.com/expressif/pecl-event-libevent

But when I run my websocket, I have an error on the line 301 of the file LibEventLoop.php:

```
$this->timerCallback = function ($_, $_, $timer) {
```
tell me that $_ is redefined. I just changed by:
```
$this->timerCallback = function ($_, $_1, $timer) {
```
and it works but I'am sure that my modification is not correct ;-)

kocoten1992:
didn't go that far :P, still scare the idea of self doing everything (self compile), I heard socket io do better than php in term of socket, will give it a try

titansoftime:
I use ratchet-php which utilizes the React EvenLoop. Upgrading to the new package solved this for me.

I checked the source and the official solution is ``$this->timerCallback = function ($_, $__, $timer) {...``

titansoftime:
I believe I have the current solution to this: https://github.com/ratchetphp/Ratchet/issues/357

titansoftime:
Although it "works" I had to eventually disable this as it would occasionally segfault.

clue:
The duplicate parameter names have been fixed via #45, so this should theoretically work **once libevent supports PHP7**.

There's ongoing effort to update libevent to support PHP7 in https://github.com/php/pecl-event-libevent/pull/2. However, the current version occasionally segfaults. This could not be observed in earlier version, so it is believed that this is a bug in libevent, rather than React.

As such, there's little we can do here.

In the meantime, you may want to check out the "event" extension. We've seen reports of people successfully using this on PHP7.

I believe this has been answered, so I'm closing this for now. Please come back with more details if this problem persists and we can reopen this :+1:

PR #62, opened 2016-11-17
Discourage libevent on PHP 7

cboden:
The libevent extension has not been updated for PHP 7 and I have seen reports of segfaults as a result.

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -7,12 +7,13 @@ class Factory
     public static function create()
     {
         // @codeCoverageIgnoreStart
-        if (function_exists('event_base_new')) {
-            return new LibEventLoop();
-        } elseif (class_exists('libev\EventLoop', false)) {
+        if (class_exists('libev\EventLoop', false)) {
             return new LibEvLoop;
         } elseif (class_exists('EventBase', false)) {
             return new ExtEventLoop;
+        } elseif (function_exists('event_base_new') && PHP_VERSION_ID < 70000) {
+            // only use ext-libevent on PHP < 7 for now
+            return new LibEventLoop();
         }
 
         return new StreamSelectLoop();
--- a/src/LibEventLoop.php
+++ b/src/LibEventLoop.php
@@ -29,6 +29,10 @@ class LibEventLoop implements LoopInterface
 
     public function __construct()
     {
+        if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
+            trigger_error('The libevent extension has not yet been updated for PHP 7, it is recommended you use a different loop', E_USER_WARNING);
+        }
+
         $this->eventBase = event_base_new();
         $this->futureTickQueue = new FutureTickQueue();
         $this->timerEvents = new SplObjectStorage();
--- a/src/LibEventLoop.php
+++ b/src/LibEventLoop.php
@@ -29,10 +29,6 @@ class LibEventLoop implements LoopInterface
 
     public function __construct()
     {
-        if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
-            trigger_error('The libevent extension has not yet been updated for PHP 7, it is recommended you use a different loop', E_USER_WARNING);
-        }
-
         $this->eventBase = event_base_new();
         $this->futureTickQueue = new FutureTickQueue();
         $this->timerEvents = new SplObjectStorage();


clue:
The docs (https://pecl.php.net/package/libevent) suggest libevent requires PHP < 6.0? How would one install libevent on PHP 7 in the first place?

Note there's also a PR that aims to add compatibility with PHP 7: https://github.com/php/pecl-event-libevent/pull/2

I'm undecided if it makes sense to explicitly discourage libevent. I'm perfectly fine with lowering its priority and/or removing it from the `Factory`, as this still leaves users the choice :+1:

cboden:
I've seen a [couple](https://github.com/ratchetphp/Ratchet/issues/463) [issues](https://github.com/reactphp/event-loop/issues/40) where users were able to install the extension on PHP 7 even though it's not officially supported yet and were complaining about segfaults. 

This PR doesn't prevent people from using the extension as it only triggers a warning and the Factory will not select a loop that segfaults.

smalot:
Any other feedback ? I'm looking for using PHP 7, so I'm too interested about this issue

clue:
> I've seen a couple issues where users were able to install the extension on PHP 7 even though it's not officially supported yet and were complaining about segfaults.

IMHO we should make this clear that we are not aware of any issues in React and this is likely an issue with the unofficial libevent bindings like so https://github.com/reactphp/event-loop/issues/55#issuecomment-271234568:

> The libevent extension does not currently support PHP7. There's ongoing effort to update libevent to support PHP7 in php/pecl-event-libevent#2. However, the current version occasionally segfaults. This could not be observed in earlier version, so it is believed that this is a bug in libevent, rather than React.
As such, there's little we can do here. Discussion about this is probably best directed at #40.

And in the same vein then actively encourage users to check out recommended extensions for PHP 7 like so:

> In the meantime, you may want to check out the "event" extension. We've seen reports of people successfully using this on PHP7.

cboden:
What's the benefit of keeping factory order unchanged? I see a negative in segfaults but I don't see a positive to keeping it as-is.

src/Factory.php
```diff
@@ -7,12 +7,12 @@ class Factory
...
         // @codeCoverageIgnoreStart
-        if (function_exists('event_base_new')) {
-            return new LibEventLoop();
-        } elseif (class_exists('libev\EventLoop', false)) {
+        if (class_exists('libev\EventLoop', false)) {
             return new LibEvLoop;
         } elseif (class_exists('EventBase', false)) {
             return new ExtEventLoop;
+        } elseif (function_exists('event_base_new') && version_compare(PHP_VERSION, '7.0.0', '<')) {
+            return new LibEventLoop();
```
clue: Makes perfect sense to lower priority here and do not pick this extension on PHP 7+ at all :+1:

src/LibEventLoop.php
```diff
@@ -31,6 +31,10 @@ class LibEventLoop implements LoopInterface
 
     public function __construct()
     {
+        if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
+            trigger_error('The libevent extension has not yet been updated for PHP 7, it is recommended you use a different loop', E_USER_WARNING);
+        }
```
clue: I'm not a big fan of this tbqh :-1: 

With the above check in place we do not automatically create the `LibEventLoop` from our `Factory`. Afaict this is what most people use, so this should resolve this for the most part?

This means that if this constructor is invoked, it's explicitly been called. IMHO this is already advanced usage and we should not trigger a warning in this case.

Otherwise, how could a user possibly ignore this warning? Using `@`-suppression would result in *zero* output if the extension can not be loaded all.

Also, what happens when the extension will be updated to support PHP 7 again? We can't really detect this until we provide an update, in which case the above warning would be superfluous.

cboden: If the user has upgraded to PHP7 and somehow kept the extension installed, from the reports I've received, it's going to segfault. The trigger_error only issues a warning, the object will still be instantiated an be able to be used. 

> Also, what happens when the extension will be updated to support PHP 7 again?

It looks like that extension isn't going to be...but if it does we'll have to quickly update this to support it. Even if this happens the compare only issues a warning.

cboden: If the consensus is still a :-1: on this I'm good to remove this check in order to get the updated Factory order into the next release.

clue: I'm not sure we had a consensus here, it was basically just my point of view :-) But if this error is removed, I'm all for getting this in! :+1:

cboden:
Updated. some > none  😀

We can discuss the warning in IRC or another issue.

src/Factory.php
```diff
@@ -7,12 +7,12 @@ class Factory
...
     {
         // @codeCoverageIgnoreStart
-        if (function_exists('event_base_new')) {
-            return new LibEventLoop();
-        } elseif (class_exists('libev\EventLoop', false)) {
+        if (class_exists('libev\EventLoop', false)) {
             return new LibEvLoop;
         } elseif (class_exists('EventBase', false)) {
             return new ExtEventLoop;
+        } elseif (function_exists('event_base_new') && version_compare(PHP_VERSION, '7.0.0', '<')) {
```
clue: Will update this to use `PHP_VERSION_ID` instead :shipit:

PR #65, opened 2017-01-09
ExtEventLoop: No longer suppress all errors

mamciek:


src/ExtEventLoop.php
```diff
@@ -178,8 +178,7 @@ public function tick()
 
         $this->futureTickQueue->tick();
 
-        // @-suppression: https://github.com/reactphp/react/pull/234#discussion-diff-7759616R226
-        @$this->eventBase->loop(EventBase::LOOP_ONCE | EventBase::LOOP_NONBLOCK);
+        $this->eventBase->loop(EventBase::LOOP_ONCE | EventBase::LOOP_NONBLOCK);
```
clue: How can we verify the above comment is no longer relevant?

Can you add some tests to verify this?

mamciek: I do not know if the above comment is relevant or not. I do know that this @-suppression supress ALL php errors that is ever generated by PHP which is very bad

clue: I understand, but if fix this now we should make sure to avoid any regressions in the future.

Can you add some tests to verify this?

jsor: I've did some tests today but i can't reproduce this. The comment in https://github.com/reactphp/react/pull/234#discussion-diff-7759616R226 is also not clear in **what** exactly produces these warnings. Maybe @nrk or @jmalloc can shed some light on this.

I agree that the `@` is really bad here because it also supresses any warnings and errors from the callbacks.

jmalloc: IIRC, the initial problem was that `EventBase::loop()` emits an (additional?) warning after an error in a user-defined callback. So, perhaps calling `trigger_error()` in a timer function, or something like that would reproduce it. I can't recall if I witnessed this myself, but I kept the @-suppression at @nrk's request.

I can see where this warning might be an issue, especially when using a strict error handler, but it's arguably the responsibility of the user to handle/suppress any such errors in callbacks. So as long as we're not producing them in the loop implementation ourselves it's probably OK to remove the `@`.

jsor: @jmalloc Thanks for the feedback!

> IIRC, the initial problem was that `EventBase::loop()` emits an (additional?) warning after an error in a user-defined callback. So, perhaps calling `trigger_error()` in a timer function, or something like that would reproduce it. I can't recall if I witnessed this myself, but I kept the @-suppression at @nrk's request.

Yes, i thought so, but couldn't reproduce it. Maybe it's fixed in the extension itself in the meantime.

I'd say it's *better* to have additional warnings than suppressing everything. Any thoughts @reactphp/core?

clue: > I'd say it's better to have additional warnings than suppressing everything.

:+1: here!

> Yes, i thought so, but couldn't reproduce it. Maybe it's fixed in the extension itself in the meantime.

If we can't reproduce this now, this means we should at least add a test this is our current assumption now (so we spot any regressions in the extension) :+1:

nrk: Probably fixed in the extension itself as I cannot reproduce it anymore. Yes @jmalloc the warning I was talking was additional and not to related to the actual error generated inside a callback, this made things very lousy. I'm absolutely :+1: to remove the @- suppression, it's a stupid hack and I hated to have it back then.

jsor: @nrk Many thanks for the info! I'd be great if you could confirm that this test would catch regressions: https://github.com/reactphp/event-loop/compare/master...jsor:ext-event-remove-error-suppression#diff-d6fed219f7ad523ad23fd345800fbf12

nrk: @jsor yes conceptually this test should do the trick. On a related note, I probably still have an old VM around at home on my NAS with an old development environment from back when I was actively contributing to reactphp (:cry:) so maybe I can try to dig it up and double check against it. I'll let you know ASAP!

jsor: @nrk That'd be great, thanks!

nrk: Alright I was able to reproduce the issue.

First of all I was slightly wrong, I was going by memory after all: using `trigger_error()` doesn't produce any additional warning, it's only exceptions thrown inside callbacks registered in the loop that trigger the following warning `PHP Warning:  EventBase::loop(): An error occurred while invoking the callback in /home/adaniele/Development/react/vendor/react/event-loop/src/ExtEventLoop.php on line 205`.

Aside from that, I wasn't initially able to reproduce it because it seems to working as expected using PHP 7.x where no additional warnings are emitted, on the other hand when using PHP 5.x these warnings are indeed emitted. This sounded kind of strange so I inspected the source code of the extension and found the culprit: errors when invoking callbacks are handled differently in [`php5/classes/event.c`](https://bitbucket.org/osmanov/pecl-event/src/41be9821b69ce996d59b750ecdbd9c07dffe192b/php7/classes/event.c?fileviewer=file-view-default#event.c-84:90) compared to [`php7/classes/event.c`](https://bitbucket.org/osmanov/pecl-event/src/41be9821b69ce996d59b750ecdbd9c07dffe192b/php5/classes/event.c?fileviewer=file-view-default#event.c-76:82), and I think that the originally intended behaviour is more akin to the one implemented for the PHP 7.x variant. I guess it's worth to ask about this to the original developer just to make sure, personally I think that the PHP 5.x variant should be fixed to implement the same logic of the PHP 7.x one.

Anyway, I also encountered a weird behaviour. Consider the following snippet:

```php
$loop = new React\EventLoop\ExtEventLoop();

$errors = array();
set_error_handler(function ($errno, $errstr) use (&$errors) {
    $errors[] = $errstr;
});

$loop->addTimer(0.001, function () {
    throw new Exception("Exception inside my callback");
});

try {
  $loop->run();
} catch (Exception $ex) {
  restore_error_handler();
}

var_dump($errors);
```

The output is:

```
array(0) {
}
```

Notices aren't printed as expected but... wait `$errors` is empty 🤔  On the other hand, running the same snippet with `set_error_handler()` commented produces the following output:

```
PHP Warning:  EventBase::loop(): An error occurred while invoking the callback in /home/adaniele/Development/react/vendor/react/event-loop/src/ExtEventLoop.php on line 205
array(0) {
}
```

😮

jsor: That is weired 😕

Interestingly, it looks like the following code hides **only** the warnings from ext-event:

```php
$loop = new React\EventLoop\ExtEventLoop();

set_error_handler(function () {
    return false; // <-- Just pass through to the standard PHP error handler
});

$loop->addTimer(0.001, function () {
    echo $undefined;
});

$loop->addTimer(0.001, function () {
    throw new Exception("Exception inside my callback");
});

$loop->run();
restore_error_handler();
```

Produces just:

```
PHP Notice:  Undefined variable: undefined in /Users/jsor/Code/~react/event-loop/test.php on line 12

PHP Fatal error:  Uncaught exception 'Exception' with message 'Exception inside my callback' in /Users/jsor/Code/~react/event-loop/test.php:16
```

Without the error handler, the ext-event warning is also shown:

```
PHP Fatal error:  Uncaught exception 'Exception' with message 'Exception inside my callback' in /Users/jsor/Code/~react/event-loop/test.php:16

PHP Warning:  EventBase::loop(): An error occurred while invoking the callback in /Users/jsor/Code/~react/event-loop/src/ExtEventLoop.php on line 191

PHP Fatal error:  Uncaught exception 'Exception' with message 'Exception inside my callback' in /Users/jsor/Code/~react/event-loop/test.php:16
```

jsor: I think we have 3 options now:

1. Keep the @-suppression
2. Use the workaround from my previous comment by temporarily set the error handler
3. Remove the @-suppression and document the change and it's impact in the changelog

My preference is option 3.

nrk: If we can get the author of pecl-event to fix the PHP 5.x variant of the extension to act just like the PHP 7.x one, then the 3rd option is definitely the best: recent versions of pecl-event are still perfectly compatible with older releases of PHP down to 5.4 and there shouldn't be any breaking change between 1.x and 2.x affecting our event loop implementation, so we could encourage users to upgrade to the latest verson of the extension by having a note in both the readme and the changelog.

In the meanwhile I can open an issue on [pecl-event](https://bitbucket.org/osmanov/pecl-event/issues?status=new&status=open) to let the author know about this.

jsor: 👍 
> In the meanwhile I can open an issue on pecl-event to let the author know about this.

That'd be great.

nrk: @jsor alright will do tomorrow first thing in the morning!

clue: > In the meanwhile I can open an issue on pecl-event to let the author know about this.

Yes, pushing this upstream is where this should be addressed in the first place! :+1: 

> I'd say it's better to have additional warnings than suppressing everything.
> […]
> 3. Remove the @-suppression and document the change and it's impact in the changelog

Absolutely agree here! :+1:

Given this only triggers when an uncaught warning or exception is passed through, I expect this to have very little impact in actual consumer code. After all, this is something that *should* never happen in the first place. Also, this only triggers an *additional* warning, so I agree this should be safe to add regardless :+1: 

This means that even if upstream takes longer to fix this and this be inconsistent across different upstream versions, I think it makes perfect sense to get this in now already :+1:

nrk: > Given this only triggers when an uncaught warning or exception is passed through, I expect this to have very little impact in actual consumer code. 

@clue the additional warning is emitted even when the exception **is** caught, see the example in [my previous comment](https://github.com/reactphp/event-loop/pull/65#discussion_r99604567).

mamciek:
The problem is half solved now :) One of the @ suppression was removed with https://github.com/reactphp/event-loop/commit/8fa040756c3a62de2e5e160c2a4d63595c0c658e

src/ExtEventLoop.php
jsor: > @jsor alright will do tomorrow first thing in the morning!

@nrk Are you still planning to file an issue in the pecl-event repo? Otherwise i can do it.

> @clue the additional warning is emitted even when the exception is caught, see the example in my previous comment.

Yes, but only if the exception bubbles up to the `$this->eventBase->loop()` call. This usually results in an uncaught exception anyway since in almost all cases `$loop->run()` is called at the end of the script.

nrk: @jsor sure that's still my plan, actually I was modifying the extension myself to test this change and eventually have a PR ready if the author agrees, just ended up having unexpected stuff going on at work so couldn't do it yet. As for uncaught exceptions: yes and I realized later that I misinterpreted what @clue meant with _passed through_.

jsor: > sure that's still my plan, actually I was modifying the extension myself to test this change and eventually have a PR ready if the author agrees, just ended up having unexpected stuff going on at work so couldn't do it yet.

No hurries, very appreciated 👍

clue:
@mamciek May I ask you to rebase this on the new "0.4" branch so we get this in for the upcoming v0.4.3 release? :shipit:

mamciek:
sure: https://github.com/reactphp/event-loop/pull/80

mamciek:
Sure. Done.

jsor:
Thanks! 👍

PR #67, opened 2017-01-11
StreamSelectLoop: Test suite uses signal constant names in data provider

martinschroeder:
Define constants `SIGUSR1`, `SIGHUP` and `SIGTERM` if they are not defined yet. The signal constants are used in data provider `StreamSelectLoopTest::signalProvider()` and cause a notice when the pcntl extension is not loaded.

```
Notice: Use of undefined constant SIGUSR1 - assumed 'SIGUSR1'
```

### DIFF
--- a/tests/StreamSelectLoopTest.php
+++ b/tests/StreamSelectLoopTest.php
@@ -40,9 +40,9 @@ class StreamSelectLoopTest extends AbstractLoopTest
     public function signalProvider()
     {
         return [
-            ['SIGUSR1', SIGUSR1],
-            ['SIGHUP', SIGHUP],
-            ['SIGTERM', SIGTERM],
+            ['SIGUSR1'],
+            ['SIGHUP'],
+            ['SIGTERM'],
         ];
     }
 
@@ -52,7 +52,7 @@ class StreamSelectLoopTest extends AbstractLoopTest
      * Test signal interrupt when no stream is attached to the loop
      * @dataProvider signalProvider
      */
-    public function testSignalInterruptNoStream($sigName, $signal)
+    public function testSignalInterruptNoStream($signal)
     {
         if (!extension_loaded('pcntl')) {
             $this->markTestSkipped('"pcntl" extension is required to run this test.');
@@ -78,7 +78,7 @@ class StreamSelectLoopTest extends AbstractLoopTest
      * Test signal interrupt when a stream is attached to the loop
      * @dataProvider signalProvider
      */
-    public function testSignalInterruptWithStream($sigName, $signal)
+    public function testSignalInterruptWithStream($signal)
     {
         if (!extension_loaded('pcntl')) {
             $this->markTestSkipped('"pcntl" extension is required to run this test.');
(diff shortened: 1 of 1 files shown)


clue:
I'm currently undecided so I won't vote yet. Can we really pass values for these missing constants or should we rather make sure the tests are skipped in this case?

martinschroeder:
The tests are skipped when pcntl is not loaded. The problem is that the constants are used in a data provider and PHPUnit will allways call the data provider before the test is skipped from within.

An alternative solution would be to do the checks in the data provider and provide defaults. This way one can get rid of notices without defining constants.

clue:
IMHO it makes more sense to make sure the data provider is only invoked if the extension is actually available. If this is not possible, then I would suggest using signal constant names only and accessing values only through the `constant()` function.

What do you think?

martinschroeder:
Using `constant()` still triggers a warning:
```
Warning: constant(): Couldn't find constant SIGUSR1
```
I changed the code to do the checks in the data provider. I am not aware of any way to prevent the data provider from being called by PHPUnit. Another solution would be to do the signal testing in another test class and do the check in the setup method but that is too much work IMHO.

jsor:
From the PHPUnit docs (second blue box):

> All data providers are executed before both the call to the setUpBeforeClass static method and the first call to the setUp method. Because of that you can't access any variables you create there from within a data provider. This is required in order for PHPUnit to be able to compute the total number of tests. 

https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers.examples.DependencyAndDataProviderCombo.php

Also related: sebastianbergmann/phpunit#836

This current solution looks like a good compromise, so i'm 👍

tests/StreamSelectLoopTest.php
```diff
@@ -51,9 +51,9 @@ public function testStopShouldPreventRunFromBlocking($timeLimit = 0.005)
     public function signalProvider()
     {
         return [
-            ['SIGUSR1', SIGUSR1],
-            ['SIGHUP', SIGHUP],
-            ['SIGTERM', SIGTERM],
+            ['SIGUSR1', defined('SIGUSR1') ? SIGUSR1 : 10],
+            ['SIGHUP', defined('SIGHUP') ? SIGHUP : 1],
+            ['SIGTERM', defined('SIGTERM') ? SIGTERM : 15],
```
clue: This looks okay to me. My suggestion would have been to only return the constant names here and then use `constant()` to access their values within the test instead. This way we would not have to provide any defaults at all and could still skip the tests (possibly via `defined()`). What do you think about this?

clue:
Ping @martinschroeder, what do you think about the above suggestion?

clue:
I've updated the base branch, can you force-push a rebased version? :+1:

martinschroeder:
I rebased to PR on the 0,4 branch. I can change the test cases to use `constant()`, but i actually think that providing fallback values within the data provider improves readability of the test cases.

martinschroeder:
I rebased the PR and changed the data provider to use signal constant names as suggested.

PR #68, opened 2017-01-11
Avoid deprecation warnings in test suite due to deprecation of getMock() in PHPUnit

martinschroeder:
The method `getMock()` was deprecated with the [release of PHPUnit 5.4](https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-5.4.0). The same release introduced `createMock()` that should be used to create mock objects instead. This PR modifies `React\Tests\EventLoop\TestCase::createCallableMock()` to use the new mock API if available.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -36,6 +36,16 @@ class TestCase extends \PHPUnit_Framework_TestCase
 
     protected function createCallableMock()
     {
-        return $this->getMock('React\Tests\EventLoop\CallableStub');
+        $stub = 'React\Tests\EventLoop\CallableStub';
+        
+        if (method_exists($this, 'createMock')) {
+            return $this->createMock($stub);
+        }
+        
+        if (method_exists($this, 'getMockBuilder')) {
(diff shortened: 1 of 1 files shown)


jsor:
We've currently decided to add PHPUnit 4.8 to require-dev of each component, see reactphp/react#354. This will fix this issue.

As a side note: i've also fixed this latetly in react/promise by using

```php
$this
    ->getMockBuilder('React\Promise\Stub\CallableStub')
    ->getMock();
```
This works in all version of PHPUnit, see https://github.com/reactphp/promise/commit/73f2304f08882c00f510a3dc74bd58ed4a28bcd6#diff-d4c8c6dc8769324fc27cfdac19f05caf

martinschroeder:
Should i change the PR to use mock builder or close it and let you introduce PHPUnit 4.8 instead?

jsor:
I'd vote for getting this in for forward compatibility and to avoid warnings when users run the tests not from vendors but with their local version of PHPUnit.

PR #72, opened 2017-02-03
Drop tick() from LoopInterface

jsor:
References: https://github.com/reactphp/event-loop/pull/52#issuecomment-239688446, async-interop/event-loop#9

For the tests, i've added a [helper function](https://github.com/reactphp/event-loop/pull/72/files#diff-d4c8c6dc8769324fc27cfdac19f05cafR44) to `TestCase`.

Closes #52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.5.0 (xxxx-xx-xx)
+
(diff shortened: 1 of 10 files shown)


jsor:
I'd like to hear from @WyriHaximus about this because IIRC he's using `tick()` in a few of his libs.

jsor:
Since this a BC break, i'd like to have also @cboden's approval here.

PR #83, opened 2017-03-15
Add support for pecl-ev

rhclayto:
This pull request adds support for the pecl-ev PHP extension.

https://bitbucket.org/osmanov/pecl-ev
http://docs.php.net/ev

I have it running locally & it seems to be working fine, as well as or better than the StreamSelectLoop. Poking & prodding is welcome of course.

### DIFF
--- a/src/ExtEvLoop.php
+++ b/src/ExtEvLoop.php
@@ -1,0 +1,210 @@
+<?php
+
+namespace React\EventLoop;
+
+use Ev;
+use EvIo;
+use EvTimer;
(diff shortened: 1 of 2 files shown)


clue:
Thanks for filing this! How does this one compare to #12 and #25?

rhclayto:
Uh, yeah, I guess i should have checked better before making the pull request. I suppose this one should be closed as the older issues do the same thing. (I guess the one difference is this one uses the default Ev lopp instead of an new EvLoop, not that it probably matters).

PR #86, opened 2017-03-25
Add tests for double watchers to be ignored

kelunik:
This tests the current behavior. I think the behavior should be changed to throw an exception instead of silently ignoring those callables or just call both callables.


kelunik:
@clue Not sure whether it should be an exception or just call all callbacks.

clue:
@kelunik Both would be valid options IMHO. Do you feel like filing an issue or preparing a PR so we can discuss this and its implications? :+1:

kelunik:
Done via #87.

PR #90, opened 2017-03-27
Restructure and improve README

jsor:
* Separate into sections and add a TOC at the beginning
* Add new sections for Install, Tests and License
* Add paragraph to Usage emphasizing on recommended program structuring using a single loop instance

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,35 +5,16 @@
 
 Event loop abstraction layer that libraries can use for evented I/O.
 
-In order for async based libraries to be interoperable, they need to use the
-same event loop. This component provides a common `LoopInterface` that any
-library can target. This allows them to be used in the same loop, with one
-single `run` call that is controlled by the user.
+**Table of Contents**
 
-In addition to the interface there are some implementations provided:
+* [Quickstart example](#quickstart-example)
+* [Usage](#usage)
+* [Loop implementations](#loop-implementations)
+* [Install](#install)
+* [Tests](#tests)
+* [License](#license)
 
-* `StreamSelectLoop`: This is the only implementation which works out of the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,7 +47,94 @@ $loop->addPeriodicTimer(5, function () {
...
-**Note:** The factory is just for convenience. It tries to pick the best
-available implementation. Libraries `SHOULD` allow the user to inject an
-instance of the loop. They `MAY` use the factory when the user did not supply
-a loop.
+## Usage
+
+In order for async based libraries to be interoperable, they need to use the
+same event loop. This component provides a common `LoopInterface` that any
+library can target. This allows them to be used in the same loop, with one
+single `run()` call that is controlled by the user.
```
clue: Should this paragraph stay at the top, right below the tag line?

jsor: Yeah, i was unsure either and moved the paragraph up and down several times while working on the patch 😃 

In the end i decided to put it under Usage because i thought it fits there and might be  easily overlooked at the beginning. My guess is, that users skip the "promotion" parts at just jump straight to the usage section.

But i'd be fine with both locations. @WyriHaximus, @cboden?

clue: The initial paragraph should try to *sell* this idea to newcomers and explain the *WHY*. This is also what will be shown as an "abstract" in many applications, e.g. the first paragraph will be shown in mobile view, while the rest will be hidden by default.

IMHO the *why* is the most important part of the documentation, but we may as well duplicate this below if this makes sense. Otherwise, we may also fine-tune this later on :+1:

jsor: I've moved the paragraph back to the top and replaced it here with a short introduction.

PR #96, opened 2017-04-05
StreamSelectLoop: Int overflow for big timer intervals

kaduev13:
### The problem

As described <a href="https://github.com/reactphp/event-loop/issues/19">here</a>, current `StreamSelectLoop` implementation restricts using time intervals more than `PHP_INT_MAX` microseconds. For 32-bit systems it's only 2148 seconds.

### Suggested solution

* Change `StreamSelectLoop` `$timeout` type from `int` (in microseconds) to `float` (in seconds);
* Replace `usleep` with `time_nanosleep`, cause `usleep` takes `int` argument in microseconds, and we can't keep more than `PHP_INT_MAX` microseconds. But `time_nanosleep` takes 2 arguments: `int` amount of seconds and `int` amount of nanoseconds, that allows `StreamSelectLoop` to handle `PHP_INT_MAX` seconds intervals;
* Extract `stream_select` and `time_nanosleep` methods calls from `StreamSelectLoop::streamSelect(...)` to test timeout values.

### DIFF
--- a/src/StreamSelectLoop.php
+++ b/src/StreamSelectLoop.php
@@ -213,18 +213,14 @@ class StreamSelectLoop implements LoopInterface
     }
 
     /**
-     * Returns integer amount of seconds in $time or null.
+     * Returns integer amount of seconds in $time.
      *
-     * @param float|null $time – time in seconds
+     * @param float $time – time in seconds
      *
-     * @return int|null
+     * @return int
      */
     private static function getSeconds($time)
     {
-        if ($time === null) {
-            return null;
-        }
-
         /*
          * Workaround for PHP int overflow:
          * (float)PHP_INT_MAX == PHP_INT_MAX => true
@@ -239,17 +235,14 @@ class StreamSelectLoop implements LoopInterface
     }
 
     /**
-     * Returns integer amount of microseconds in $time or null.
+     * Returns integer amount of microseconds in $time.
      *
-     * @param float|null $time – time in seconds
+     * @param float $time – time in seconds
      *
-     * @return int|null
+     * @return int
      */
     private static function getMicroseconds($time)
     {
-        if ($time === null) {
-            return null;
-        }
         $fractional = fmod($time, 1);
         $microseconds = round($fractional * self::MICROSECONDS_PER_SECOND);
 
@@ -257,19 +250,15 @@ class StreamSelectLoop implements LoopInterface
     }
 
     /**
-     * Returns integer amount of nanoseconds in $time or null.
+     * Returns integer amount of nanoseconds in $time.
      * The precision is 1 microsecond.
      *
-     * @param float|null $time – time in seconds
+     * @param float $time – time in seconds
      *
-     * @return int|null
+     * @return int
      */
     private static function getNanoseconds($time)
     {
-        if ($time === null) {
-            return null;
-        }
-
         return intval(self::getMicroseconds($time) * self::NANOSECONDS_PER_MICROSECOND);
     }
 
@@ -286,9 +275,9 @@ class StreamSelectLoop implements LoopInterface
      */
     protected function streamSelect(array &$read, array &$write, $timeout)
     {
-        $seconds = self::getSeconds($timeout);
-        $microseconds = self::getMicroseconds($timeout);
-        $nanoseconds = self::getNanoseconds($timeout);
+        $seconds = $timeout === null ? null : self::getSeconds($timeout);
+        $microseconds = $timeout === null ? 0 : self::getMicroseconds($timeout);
+        $nanoseconds = $timeout === null ? 0 : self::getNanoseconds($timeout);
 
         if ($read || $write) {
             $except = [];
@@ -296,7 +285,9 @@ class StreamSelectLoop implements LoopInterface
             return $this->doSelectStream($read, $write, $except, $seconds, $microseconds);
         }
 
-        $this->sleep($seconds, $nanoseconds);
+        if ($timeout !== null) {
+            $this->sleep($seconds, $nanoseconds);
+        }
 
         return 0;
     }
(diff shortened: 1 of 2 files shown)


src/StreamSelectLoop.php
```diff
@@ -220,27 +213,117 @@ private function waitForStreamActivity($timeout)
...
+     * Sleeps for $seconds and $nanoseconds.
+     *
+     * @param int|null $seconds
+     * @param int|null $nanoseconds
+     */
+    protected function sleep($seconds, $nanoseconds = 0)
+    {
+        if ($seconds > 0 || $nanoseconds > 0) {
+            time_nanosleep($seconds, $nanoseconds);
+        }
```
jsor: I'd like to see explicit handling of `$seconds`or `$nanoseconds`being `null` here.

src/StreamSelectLoop.php
```diff
@@ -220,27 +213,117 @@ private function waitForStreamActivity($timeout)
...
+         * Workaround for PHP int overflow:
+         * (float)PHP_INT_MAX == PHP_INT_MAX => true
+         * (int)(float)PHP_INT_MAX == PHP_INT_MAX => false
+         * (int)(float)PHP_INT_MAX == PHP_INT_MIN => true
+         */
+        if ($time == PHP_INT_MAX) {
+            return PHP_INT_MAX;
+        }
+
+        return $time === null ? null : intval(floor($time));
```
jsor: Minor nitpicking: for consistency with the other 2 methods, i'd like to an early return for `$time === null`:

```php
if ($time === null) {
    return null;
}

// ...

return intval(floor($time));
```

src/StreamSelectLoop.php
```diff
@@ -220,27 +213,124 @@ private function waitForStreamActivity($timeout)
...
+     * Returns integer amount of microseconds in $time or null.
+     *
+     * @param float|null $time – time in seconds
+     *
+     * @return int|null
+     */
+    private static function getMicroseconds($time)
+    {
+        if ($time === null) {
+            return null;
```
jsor: Should probably return `0`. `stream_select` accepts `null` only for the `$seconds` parameter.

src/StreamSelectLoop.php
```diff
@@ -220,27 +213,124 @@ private function waitForStreamActivity($timeout)
...
+     * The precision is 1 microsecond.
+     *
+     * @param float|null $time – time in seconds
+     *
+     * @return int|null
+     */
+    private static function getNanoseconds($time)
+    {
+        if ($time === null) {
+            return null;
```
jsor: Should probably return `0`, see above.

src/StreamSelectLoop.php
```diff
@@ -220,27 +213,124 @@ private function waitForStreamActivity($timeout)
...
     }
+
+    /**
+     * Proxy for built-in stream_select method.
+     *
+     * @param array $read
+     * @param array $write
+     * @param array $except
+     * @param int|null $seconds
+     * @param int|null $microseconds
```
jsor: Should only accept `int`, see also above.

kaduev13:
@jsor, thank you for the review. I think that previous implementation was not explicit, cause `$timeout` checking for `null` value was duplicated in 4 places (`getSeconds`, `getMicroseconds`, `getNanoseconds` and `sleep`). I changed this behavior in c78624d: I removed all unnecessary `null` checks and moved them inside `streamSelect` method. Also, I fixed the things you pointed ;)

src/StreamSelectLoop.php
```diff
@@ -220,27 +213,112 @@ private function waitForStreamActivity($timeout)
...
+     * @param float|null   $timeout Activity timeout in seconds, or null to wait forever.
      *
      * @return integer|false The total number of streams that are ready for read/write.
      * Can return false if stream_select() is interrupted by a signal.
      */
     protected function streamSelect(array &$read, array &$write, $timeout)
     {
+        $seconds = $timeout === null ? null : self::getSeconds($timeout);
+        $microseconds = $timeout === null ? 0 : self::getMicroseconds($timeout);
+        $nanoseconds = $timeout === null ? 0 : self::getNanoseconds($timeout);
```
jsor: Calculation of `$microseconds` and `$nanoseconds` can probably be moved into the `if`s so they are only calculated when needed.

src/StreamSelectLoop.php
```diff
@@ -220,27 +213,112 @@ private function waitForStreamActivity($timeout)
...
+     */
+    private static function getSeconds($time)
+    {
+        /*
+         * Workaround for PHP int overflow:
+         * (float)PHP_INT_MAX == PHP_INT_MAX => true
+         * (int)(float)PHP_INT_MAX == PHP_INT_MAX => false
+         * (int)(float)PHP_INT_MAX == PHP_INT_MIN => true
+         */
+        if ($time == PHP_INT_MAX) {
```
jsor: Strict comparision (`===`)?

kaduev13: @jsor, `$time` is float, so strict comparison will always return `false`.

```php
(float) PHP_INT_MAX === PHP_INT_MAX; //is false
(float) PHP_INT_MAX == PHP_INT_MAX; //is true
```

jsor: 👍 (maybe add a comment that loose comparision is intentional)

src/StreamSelectLoop.php
kaduev13: @jsor, done in bb81383 😃

src/StreamSelectLoop.php
kaduev13: @jsor I added more description about this tricky  `if` in 3c8a5d6 😄

jsor:
ping @clue

kaduev13:
@clue, can you please describe, why did you mark this PR as BC Break? Imo, only internal representation of time changed and no public interfaces modified...

src/StreamSelectLoop.php
```diff
@@ -220,27 +213,117 @@ private function waitForStreamActivity($timeout)
...
+    }
+
+    /**
      * Emulate a stream_select() implementation that does not break when passed
      * empty stream arrays.
      *
      * @param array        &$read   An array of read streams to select upon.
      * @param array        &$write  An array of write streams to select upon.
-     * @param integer|null $timeout Activity timeout in microseconds, or null to wait forever.
+     * @param float|null   $timeout Activity timeout in seconds, or null to wait forever.
```
clue: This is a rather subtle BC break as consumers *could* possibly rely on this `protected` API. Not saying this is an issue with your proposed change, it's more that we ~screwed up~ could have done better with defining our external API.

PR #110, opened 2017-10-13
Documentation for advanced stream API

### LINKED ISSUE #87: Allow multiple IO callbacks or throw
Currently, multiple IO callbacks are ignored, only the first one will work. #86 fixes this to be consistent, but silently ignoring should be changed to either throw exceptions or call all handlers and not ignore them.

clue:
The stream API is currently mostly undocumented. This PR adds documentation for the existing stream API and then subsequently removes the unneeded and undocumented loop argument that was previously passed to stream listeners.

This is a subtle BC break. Empirical evidence (including our examples, tests and other components) suggest that most consumers will not be affected by this. The added documentation ensures that we now follow a strict API and will not introduce a BC break in the future.

Trying to create documentation for this API is not exactly trivial, as it's very low-level and has existed in this form almost since its inception. Given that most consumers SHOULD NOT use this API at all, I've marked the stream API as advanced and linked to the Stream component instead.

Performance improvement is not a major motivation here, but shows a negligible improvement anyway (running examples 92 and 94).

If you want to review, consider also looking at the individual commits.

Builds on top of #100 and #102.
Resolves / closes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,15 +358,24 @@ the same time is not guaranteed.
 The `removeReadStream(resource $stream): void` method can be used to
 remove the read event listener for the given stream.
 
+Removing a stream from the loop that has already been removed or trying
+to remove a stream that was never added or is invalid has no effect.
+
 ### removeWriteStream()
 
 The `removeWriteStream(resource $stream): void` method can be used to
 remove the write event listener for the given stream.
 
+Removing a stream from the loop that has already been removed or trying
+to remove a stream that was never added or is invalid has no effect.
+
 ### removeStream()
 
 The `removeStream(resource $stream): void` method can be used to
 remove all listeners for the given stream.
+
+Removing a stream from the loop that has already been removed or trying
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -291,6 +292,116 @@ echo 'a';
...
+> Advanced! Note that this low-level API is considered advanced usage.
+  Most use cases should probably use the higher-level
+  [writable Stream API](https://github.com/reactphp/stream#writablestreaminterface)
+  instead.
+
+The `addWriteStream(resource $stream, callable $callback): void` method can be used to
+register a listener to be notified when a stream is ready to write.
+
+The first parameter MUST be a valid stream resource that supports
+checking whether it is ready to read by this loop implementation.
```
jsor: ...it is ready to **_write_** by this...

clue: Thanks for spotting, fixed and amended! :shipit:

PR #118, opened 2017-11-10
[RFC] Remove removeStream() method, use removeReadStream/removeWriteStream instead

clue:
The `removeStream()` acts as a shortcut to essentially call both `removeReadStream()` and `removeWriteStream()` in one go. It's unclear how much value this method actually provides and it looks like this has attracted some low quality code in the past. As such, I'm filing this PR as an RFC to see if it makes sense to drop this unneeded method.

Refs #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ $loop->addReadStream($server, function ($server) use ($loop) {
         $written = fwrite($conn, $data);
         if ($written === strlen($data)) {
             fclose($conn);
-            $loop->removeStream($conn);
+            $loop->removeWriteStream($conn);
         } else {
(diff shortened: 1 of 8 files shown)


cboden:
Seems logical. reactphp/stream and reactphp/zmq will require updating.

clue:
> Seems logical. reactphp/stream and reactphp/zmq will require updating.

Yep, here's the first PR https://github.com/reactphp/stream/pull/121 :shipit:

edusalguero:
Hi!
About:
> Seems logical. reactphp/stream and reactphp/zmq will require updating.

Here my PR to update reactphp/zmq#53

PR #123, opened 2017-12-01
Fix ExtEventLoop to keep track of stream resources (refcount)

clue:
It has come to our attention that certain loop implementations do not keep track of assigned stream resources (refcount). This means that stream resources may be garbage collected despite being added to the loop when leaving the scope where they have been defined. Arguably, this is a design decision that makes sense for a number of PHP extension modules.

This issue is implicitly avoiding by using react/stream, which stores the stream resource in a member variable and thus prevents garbage collection. This means that this is not an issue for most people that rely on this component. However, this issue can be exhibited quite easily by using the EventLoop directly.

Given that our loop interface tries to abstract these underlying differences, we may have to keep track of stream resources within our loop abstractions.

~~~I'm marking this as WIP until we can pinpoint, reproduce and fix the underlying issue or work out another reasonable work around.~~~

### DIFF
--- a/tests/AbstractLoopTest.php
+++ b/tests/AbstractLoopTest.php
@@ -61,6 +61,39 @@ abstract class AbstractLoopTest extends TestCase
         $this->tickLoop($this->loop);
     }
 
+    public function testAddReadStreamReceivesDataFromStreamReference()
+    {
+        $this->received = '';
+        $this->subAddReadStreamReceivesDataFromStreamReference();
+        $this->assertEquals('', $this->received);
+
+        $this->assertRunFasterThan($this->tickTimeout * 2);
+        $this->assertEquals('[hello]X', $this->received);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


clue:
Removed the WIP marker as this PR now contains a minimal work around and we can no longer reproduce this issue with this fix applied (see also Travis output) :shipit:

tests/AbstractLoopTest.php
```diff
@@ -61,6 +61,39 @@ public function testAddReadStreamIgnoresSecondCallable()
...
+        $this->received = '';
+        $this->subAddReadStreamReceivesDataFromStreamReference();
+        $this->assertEquals('', $this->received);
+
+        $this->assertRunFasterThan($this->tickTimeout * 2);
+        $this->assertEquals('[hello]X', $this->received);
+    }
+
+    /**
+     * Telper for above test. This happens in another helper method to verify
```
andig: Helper typo

PR #124, opened 2017-12-01
Test against PHP 7.2

carusogabriel:
With the release of [`PHP 7.2`](http://php.net/archive/2017.php#id2017-11-30-1), would be interesting to test it.


clue:
The test currently fail because the required extensions can not be installed on PHP 7.2. Can you look into updating the `travis-init.sh`?

carusogabriel:
@clue Done!

PR #133, opened 2017-12-04
Remove unneeded isTimerActive() to reduce API surface

clue:
Builds on top of #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,6 @@ For the code of the current stable 0.4.x release, checkout the
     * [addTimer()](#addtimer)
     * [addPeriodicTimer()](#addperiodictimer)
     * [cancelTimer()](#canceltimer)
-    * [isTimerActive()](#istimeractive)
     * [futureTick()](#futuretick)
     * [addSignal()](#addsignal)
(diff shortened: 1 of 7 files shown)


clue:
@dearkey Good question, can you gist your use case and the problem you're trying to solve? During our discussions for this ticket and #102 we didn't come up with any use cases for this API, so we decided to remove it as part of the API cleanup. Likewise, this ticket hasn't been referenced for more than a year.

dearkey:
i use now a global variable to set the active state. when i use a addPeriodicTimer to start an other one i have to check if its already running. because the second will be stopped when nothing is to do.

clue:
@dearkey The way I understand this, this sounds like a reasonable approach :+1:

Using an additional variable to hold this application state is a pattern we see in a number of applications (e.g. https://github.com/clue/reactphp-ssh-proxy/blob/c1732bad827672d644ea28f4806e08b86271608a/src/SshSocksConnector.php#L168-L171 and many others) and it's my understanding this solves this problem much better than exposing this state in this lower-level component :+1:

PR #141, opened 2017-12-20
[RFC] Remove all callable type hints for consistency and performance reasons

clue:
I'm filing this PR as an RFC to see if it makes sense to drop these type hints.

### DIFF
--- a/src/ExtEventLoop.php
+++ b/src/ExtEventLoop.php
@@ -67,7 +67,7 @@ final class ExtEventLoop implements LoopInterface
         $this->createStreamCallback();
     }
 
-    public function addReadStream($stream, callable $listener)
+    public function addReadStream($stream, $listener)
     {
         $key = (int) $stream;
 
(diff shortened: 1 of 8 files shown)


clue:
I've seen a 5% to 10% performance improvement running any of the benchmarks from the `examples/`folder and have also seen some similar performance improvements in running some higher level applications (I/O intensive) with this applied.

Note that I'm not suggesting we remove *all* type hints. This PR merely aims to remove the `callable` type hint only – also for consistency reasons with the whole API. Also note that this does not contradict our efforts to eventually update to PHP 7 only (https://github.com/reactphp/react/issues/374), in which case we will likely reconsider adding consistent type hints for all relevant APIs.

kelunik:
I can't reproduce any improvements in `examples/94-benchmark-timers-delay.php` either. Which benchmark specifically did you run and which numbers did you get?

PR #149, opened 2018-02-06
Properly clean up event watchers for ext-event and ext-libev

clue:
Also refs #104

### DIFF
--- a/src/ExtLibevLoop.php
+++ b/src/ExtLibevLoop.php
@@ -55,6 +55,7 @@ final class ExtLibevLoop implements LoopInterface
             },
             function ($signal) {
                 if ($this->signals->count($signal) === 0) {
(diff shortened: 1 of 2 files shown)


src/ExtLibevLoop.php
```diff
@@ -110,6 +112,7 @@ public function removeWriteStream($stream)
 
         if (isset($this->writeEvents[$key])) {
             $this->writeEvents[$key]->stop();
+            $this->loop->remove($this->readEvents[$key]);
```
jsor: Looks like this line causes build failures in PHP < 7, eg. https://travis-ci.org/reactphp/event-loop/jobs/337955988#L1463

clue: Thanks for spotting, fixed the copy-pasta 🍝  :shipit:

PR #150, opened 2018-02-06
Refactor and simplify signal handling logic

clue:
This PR refactors some of the internal signal handling logic. It avoids a number of circular references to avoid relying on GC collecting cycles. It does not affect the public API, this is internal only.

Builds on top of #104 and #149


clue:
Removed WIP marker, this is now ready for review :shipit:

PR #151, opened 2018-02-06
[RFC] Improve compatibility with legacy versions

clue:
Because _why not_… :-)

Now more seriously: This project is a low level lib that is used as a building block for quite a few higher level abstractions on top of it. As such, compatibility (even with significantly outdated versions) is a major concern to me.

Note that I'm not suggesting putting _significant amount_ of work into this. The patch is already here and, personally, I see little harm in supporting this.

Also note that I'm not suggesting we need to keep support indefinitely. Should this ever turn out to be a burden in the future, e.g. because we actually _require_ any new language features or some external lib, then I'm all for dropping support again. If you want to discuss this further, I would suggest directing this to https://github.com/reactphp/react/issues/374 instead 👍

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: php
 
 php:
+# - 5.3 # requires old distro, see below
   - 5.4
   - 5.5
   - 5.6
@@ -13,6 +14,9 @@ php:
 dist: trusty
 
 matrix:
+  include:
+    - php: 5.3
+      dist: precise
   allow_failures:
     - php: hhvm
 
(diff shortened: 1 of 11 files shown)


nick4fake:
Improving compatibility with old versions is totally ok if that does not limit your future changes. Are you sure PHP5.3/5.4/5.5 are really important to support?

clue:
@nick4fake I think we all agree that this is a highly political topic and I believe most of this has been discussed in https://github.com/reactphp/react/issues/374#issuecomment-363553089 (another others) already.

The gist here is that *using* outdated versions is most likely a bad idea – however, older versions do no currently block or limit us on what is planned for the stable v1.0.0 release. As such, there are no immediate plans to drop support for any older versions, which allows consumers of our packages to safely update without having to worry about this for now.

In case anybody is still wondering: We highly recommend running a supported version of PHP (which again is explicitly documented in each components' install instructions).

PR #164, opened 2018-04-21
StreamSelectLoop: Improve memory consumption and runtime performance for timers

clue:
While debugging some very *odd memory issues* in a live application, I noticed that the `StreamSelectLoop` does not immediately free all related memory when a timer is cancelled. Let's not call this a "memory leak", because memory was *eventually* freed, but this clearly caused some unexpected and significant memory growth.

In many common programs, adding timeouts to operations is a very fundamental operation. As such, it's common to end up with hundred or thousands of timers which most of the time get cancelled very frequently as a successful execution will quickly cancel outstanding timeouts.

I've used the following script to demonstrate unreasonable memory growth:

```
<?php

use React\EventLoop\Factory;

require __DIR__ . '/../vendor/autoload.php';

$loop = Factory::create();

$loop->addPeriodicTimer(0.00001, function () use ($loop) {
    $timer = $loop->addTimer(60, function () { });
    $loop->cancelTimer($timer);
});

$loop->addPeriodicTimer(1.0, function () use ($loop) {
    echo memory_get_usage() . PHP_EOL;
});

$loop->run();
```

Running this on the current master branch, this peaked at around 320 MB of memory on my system (YMMV). After applying this patch, this script reports a constant memory consumption of around 0.7 MB.

The original implementation internally relied on a `SplPriorityQueue` for fast insertions of new timers, but this class lacks access to actually remove timers. This means that they would actually stay in the queue until they were originally supposed to fire.

This was apparently done for performance reasons, so this patch includes some thorough performance tests. The result here is that adding a large number of timers at the same time shows a significant performance improvement (`time php examples/92-benchmark-timers.php 20000` from 15s down to 2s). Waiting for a large number of timers to fire one after another does not show a noticeable impact (`time php examples/94-benchmark-timers-delay.php 20000` both at around 2.5s).

### DIFF
--- a/src/Timer/Timers.php
+++ b/src/Timer/Timers.php
@@ -17,6 +17,7 @@ final class Timers
     private $time;
     private $timers = array();
     private $schedule = array();
+    private $sorted = true;
 
     public function updateTime()
     {
@@ -33,7 +34,7 @@ final class Timers
         $id = spl_object_hash($timer);
         $this->timers[$id] = $timer;
         $this->schedule[$id] = $timer->getInterval() + microtime(true);
-        asort($this->schedule);
+        $this->sorted = false;
     }
 
     public function contains(TimerInterface $timer)
@@ -49,6 +50,12 @@ final class Timers
 
     public function getFirst()
     {
+        // ensure timers are sorted to simply accessing next (first) one
+        if (!$this->sorted) {
+            $this->sorted = true;
+            asort($this->schedule);
+        }
+
         return reset($this->schedule);
     }
 
@@ -59,6 +66,12 @@ final class Timers
 
     public function tick()
     {
+        // ensure timers are sorted so we can execute in order
+        if (!$this->sorted) {
+            $this->sorted = true;
+            asort($this->schedule);
+        }
+
         $time = $this->updateTime();
 
         foreach ($this->schedule as $id => $scheduled) {
@@ -78,7 +91,7 @@ final class Timers
             // re-schedule if this is a periodic timer and it has not been cancelled explicitly already
             if ($timer->isPeriodic() && isset($this->timers[$id])) {
                 $this->schedule[$id] = $timer->getInterval() + $time;
-                asort($this->schedule);
+                $this->sorted = false;
             } else {
                 unset($this->timers[$id], $this->schedule[$id]);
             }
--- a/src/Timer/Timers.php
+++ b/src/Timer/Timers.php
@@ -3,8 +3,6 @@
 namespace React\EventLoop\Timer;
 
 use React\EventLoop\TimerInterface;
-use SplObjectStorage;
-use SplPriorityQueue;
 
 /**
  * A scheduler implementation that can hold multiple timer instances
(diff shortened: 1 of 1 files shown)


src/Timer/Timers.php
```diff
@@ -77,32 +66,34 @@ public function isEmpty()
...
     public function tick()
     {
-        $time = $this->updateTime();
-        $timers = $this->timers;
-        $scheduler = $this->scheduler;
-
-        while (!$scheduler->isEmpty()) {
-            $timer = $scheduler->top();
+        // ensure timers are sorted so we can execute in order
+        if (!$this->sorted) {
```
nawarian: You have the exact same logic at line 54. You could create a private method `ensureScheduleIsSorted()` or something similar in order to do this checking and perform the sorting operation.

clue: Fair point! This is the very core of the event loop, so this is the critical code path and I would rather avoid any additional function calls. What do you think about this? :+1:

src/Timer/Timers.php
```diff
@@ -77,32 +66,34 @@ public function isEmpty()
...
+            $this->sorted = true;
+            asort($this->schedule);
+        }
 
-            if (!isset($timers[$timer])) {
-                $scheduler->extract();
-                $timers->detach($timer);
+        $time = $this->updateTime();
+        $schedule = $this->schedule;
```
nawarian: As `$this->schedule` is an array, by doing `$schedule = $this->schedule` you create a copy which looks like a waste of memory in order "to look good" at line 78.
If you use `foreach ($this->schedule as $id => $scheduled) {}` instead it should be fine (even if the list increases or decreases in number, since the iterator will operate with a local copy of it during the first loop).

enumag:
SPL data structures are so [badly written](https://medium.com/@rtheunissen/efficient-data-structures-for-php-7-9dda7af674cd) that I'm not at all surprised that their performance is not good. But I'd really like to see whether the new [Data Structures](http://php.net/manual/en/book.ds.php) would be better. In theory they should be faster than arrays since arrays have too many features to be efficient. The new Data Structures have 2 implementations. One as PHP extension and another as pure PHP composer package (useful as fallback in case you can't install the extension).

src/Timer/Timers.php
```diff
@@ -77,32 +66,34 @@ public function isEmpty()
...
 
-            if ($timers[$timer] >= $time) {
-                break;
+            // skip any timers that are removed while we process the current schedule
+            if (!isset($this->schedule[$id]) || $this->schedule[$id] !== $scheduled) {
+                continue;
             }
 
-            $scheduler->extract();
+            $timer = $this->timers[$id];
```
nawarian: Is it possible that this line will throw a notice of undefined index?

src/Timer/Timers.php
clue: Good catch! This was a leftover from the original version and it's indeed safe to remove this :+1: Updated as per your suggestion, thank you.

src/Timer/Timers.php
clue: I can see where you're coming from. This is not possible as both arrays are always modified pair-wise and as such the element is known to exist due to the check immediately before this assignment. I've added a comment to the code block below as this array may change *after* the timer has been invoked. Both cases are covered by the test suite, this code has 100% coverage.

clue:
@nawarian Thank you for the review! I've addressed all things you've pointed out and believe this is ready to be merged :shipit: :+1: 

@enumag While I don't agree with this assessment about SPL, I very much appreciate the discussion about DS! I think it may make sense to look into this for a future version, perhaps you or somebody else feels like filing a PR for this in the future so we can evaluate this? :+1: One of the design goals of this component for this component is that is *runs everywhere* and is standalone without any external dependencies, so I'm curious how this would work out eventually.

PR #169, opened 2018-06-25
Simplify test bootstrap by using dev autoloader

lcobucci:
This is already being done in other repos of the organisation, let's make it more consistent then =)


lcobucci:
These failures seem like flaky tests...

PR #219, opened 2020-12-31
Update Timer.php

cobolfoo:
Allow user data to be passed to the callback function. This is a much needed feature for complex projects (like the one I am working on) :+1:

### DIFF
--- a/src/Timer/Timer.php
+++ b/src/Timer/Timer.php
@@ -2,6 +2,7 @@
 
 namespace React\EventLoop\Timer;
 
+
 use React\EventLoop\TimerInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


clue:
@cobolfoo Thanks for filing this PR!

Are you aware of https://github.com/reactphp/event-loop/pull/102? There used to be a `setData()` method that has been removed in favor of binding arbitrary variables to closures.

From my personal experience, this is something has worked much better than assigning variables through a single "data" variable. I've build dozens of complex and simple projects with this and feel comfortable saying that I very much prefer this way and haven't encountered any issues doing so.

That being said, I'm open for discussion and I'd like to learn more about the problem you're seeing. Perhaps you can gist why you think this could be beneficial?

clue:
I'm closing this for now as it hasn't received any input in a while and I believe this has been answered. Please come back with more details if this persists and we can reopen this :+1:

PR #231, opened 2021-06-30
Improve documentation regarding deprecated Factory

clue:
This changeset improves documentation regarding deprecated Factory. As of v1.2.0 the `Loop::get()` method should be used instead.

Builds on top of #226

PR #232, opened 2021-06-30
Automatically run Loop at end of program (autorun)

clue:
When using the `Loop` class, it will automatically execute the loop at the end of
the program. This means the following example will schedule a timer and will
automatically execute the program until the timer event fires:

```php
use React\EventLoop\Loop;

Loop::addTimer(1.0, function () {
    echo 'Hello' . PHP_EOL;
});
```

As of `v1.2.0`, we highly recommend using the `Loop` class this way and omitting any
explicit [`run()`](#run) calls. For BC reasons, the explicit [`run()`](#run)
method is still valid and may still be useful in some applications, especially
for a transition period towards the more concise style.

If you don't want the `Loop` to run automatically, you can either explicitly
[`run()`](#run) or [`stop()`](#stop) it. This can be useful if you're using
a global exception handler like this:

```php
use React\EventLoop\Loop;

Loop::addTimer(10.0, function () {
    echo 'Never happens';
});

set_exception_handler(function (Throwable $e) {
    echo 'Error: ' . $e->getMessage() . PHP_EOL;
    Loop::stop();
});

throw new RuntimeException('Demo');
```

This is a pure feature addition that works across all loops and does not affect BC.

Builds on top of #226, #229 and #231

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,25 @@ explicit [`run()`](#run) calls. For BC reasons, the explicit [`run()`](#run)
 method is still valid and may still be useful in some applications, especially
 for a transition period towards the more concise style.
 
+If you don't want the `Loop` to run automatically, you can either explicitly
+[`run()`](#run) or [`stop()`](#stop) it. This can be useful if you're using
+a global exception handler like this:
+
+```php
+use React\EventLoop\Loop;
+
+Loop::addTimer(10.0, function () {
+    echo 'Never happens';
+});
+
+set_exception_handler(function (Throwable $e) {
+    echo 'Error: ' . $e->getMessage() . PHP_EOL;
+    Loop::stop();
+});
+
+throw new RuntimeException('Demo');
+```
+
 #### get()
 
 The `get(): LoopInterface` method can be used to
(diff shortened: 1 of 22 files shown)


kelunik (commented):
We've previously briefly discussed this, and I'd recommend against introducing this. Application registered shutdown functions won't be run if they're registered inside event loop watchers and a fatal error or explicit `exit;` happen, see e.g. https://3v4l.org/duM4S.

clue:
> We've previously briefly discussed this, and I'd recommend against introducing this. Application registered shutdown functions won't be run if they're registered inside event loop watchers and a fatal error or explicit `exit;` happen, see e.g. https://3v4l.org/duM4S.

The autorun functionality is a new feature that will only be used when switching to the `Loop` class. Existing code explicitly using the `$loop->run()` call or new code explicitly switching to `Loop::run()` can still explicitly run the loop without resorting to the autorun functionality. If this is a concern for your use case, the loop would not be executed from within a shutdown function in this case.

Other than that, I don't see how this would affect real-world use cases in any negative way. I've added a bunch of test cases that verify common application setups work as expected (either using the new style or the existing style). If you feel anything's missing, I'm happy to look into this and provide additional test cases :+1:

clue:
Rebased on `master` to resolve merge conflict now that #231 is in :shipit:

WyriHaximus (approved):

PR #244, opened 2022-02-19
Remove duplicate code & make code reusable

Faisal50x:


SimonFrings:
Hey @Faisal50x, thanks for bringing this up :+1:

The same topic came up in reactphp/event-loop#164 when this code was introduced. @clue did a good job at explaining that this is the very core of the event loop, thus a critical code path. Every additional function call takes up some time and we rather want to avoid this where we can.

Don't get me wrong, you're suggested changes make total sense but it's also important to look at the benefit we're gaining from this, because this is an internal adjustment and only "visible" to contributors when working on this.

What do you think about this?

Faisal50x:
@SimonFrings 
Thank you for clarifying me. I close this PR.

PR #250, opened 2022-03-16
Update supported PHP versions for loop extensions

SimonFrings:
I checked and updated the supported PHP versions with the different loop extensions.

Builds on top of #238.


clue (approved):
@SimonFrings Thanks for looking into this, changes LGTM! :shipit: 

Good to have the docs and test suite updated after #230 :+1:

WyriHaximus (approved):

PR #260, opened 2022-09-17
Set up PHPStan on GitHub Actions

nhedger:
This PR sets up PHPStan to run on GitHub Actions, as discussed in [discussions#469](https://github.com/orgs/reactphp/discussions/469).

### Overview

- [x] Sets up PHPStan to run on GitHub Actions on PHP 8.1 only
- [x] Configures PHPStan to run the analysis on the `examples`, `src` and `tests` folders
- [x] Generates the [baseline](https://phpstan.org/user-guide/baseline) so that static analysis passes immediately 

### Baseline

Because this PR aims to set up PHPStan and not address the errors it reports, I've generated a baseline to make the pipeline succeed. We'll then be able to incrementally fix the problems in future PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,3 +113,23 @@ jobs:
       - run: composer self-update --2.2 # downgrade Composer for HHVM
       - run: hhvm $(which composer) install
       - run: hhvm vendor/bin/phpunit
+
+  PHPStan:
+    name: PHPStan
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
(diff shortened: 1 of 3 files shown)


clue:
For the reference: It looks like this wasn't closed intentionally (as discussed in https://github.com/reactphp/async/pull/64). We've recently started applying PHPStan on max level into our projects with https://github.com/reactphp/async/pull/76 and we'll go ahead with this and apply PHPStan on max level to all our projects in the future.

Thank you very much for your contribution and sparking this discussion! :+1:

PR #266, opened 2023-05-04
Improve performance of `Loop` by avoiding unneeded method calls

clue:
This changeset improves performance of `Loop` slightly by avoiding unneeded method calls. Not the most significant improvement, but given the small changeset and how we plan to make the default `Loop` more prominent in ReactPHP v3, I figured this change makes sense in either case:

```bash
$ time php examples/91-benchmark-ticks.php 1000000
// old 0.83
// new 0.76
```

Builds on top of #245 and #229
See also https://github.com/orgs/reactphp/discussions/481

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -8,7 +8,7 @@ namespace React\EventLoop;
 final class Loop
 {
     /**
-     * @var LoopInterface
+     * @var ?LoopInterface
      */
     private static $instance;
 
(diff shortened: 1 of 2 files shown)


clue:
Updated to work around all legacy PHP and Windows quirks, the result is a somewhat less compact changeset, but still shows the same performance improvements. One more reason to drop legacy support in upcoming ReactPHP v3 (https://github.com/orgs/reactphp/discussions/481) to simplify this significantly! This is ready for review. :shipit:

WyriHaximus (approved):
Not as pretty as we discussed earlier this week, but a good step to improve performance for this :+1:

SimonFrings (approved):

PR #267, opened 2023-08-30
Improve performance by using `spl_object_id()` on PHP 7.2+

samsonasik:
`spl_object_id()` is exists since php 7.2, I think `spl_object_id()` can be used when php version >= 7.2

Refs:

- https://github.com/php/php-src/pull/2611
- https://developer.wordpress.com/2023/08/24/speedier-php-execution-in-wordpress-6-3/

Same implementation in reactphp/http:

https://github.com/reactphp/http/blob/c6321978bfb82de979fe4dc28eb0c08bc34937ad/src/Io/RequestHeaderParser.php#L165

**Benchmark**

**Before** **_call in loop took: 13.639ms_**

https://3v4l.org/4rUL4

**After**  **_call in loop took: 4.835ms_**

https://3v4l.org/jKtO6

### DIFF
--- a/src/Timer/Timers.php
+++ b/src/Timer/Timers.php
@@ -38,7 +38,7 @@ final class Timers
 
     public function add(TimerInterface $timer)
     {
-        $id = \spl_object_hash($timer);
+        $id = \PHP_VERSION_ID < 70200 ? \spl_object_hash($timer) : \spl_object_id($timer);
         $this->timers[$id] = $timer;
         $this->schedule[$id] = $timer->getInterval() + $this->updateTime();
         $this->sorted = false;
@@ -46,12 +46,13 @@ final class Timers
 
     public function contains(TimerInterface $timer)
     {
-        return isset($this->timers[\spl_object_hash($timer)]);
+        $id = \PHP_VERSION_ID < 70200 ? \spl_object_hash($timer) : \spl_object_id($timer);
+        return isset($this->timers[$id]);
     }
 
     public function cancel(TimerInterface $timer)
     {
-        $id = \spl_object_hash($timer);
+        $id = \PHP_VERSION_ID < 70200 ? \spl_object_hash($timer) : \spl_object_id($timer);
         unset($this->timers[$id], $this->schedule[$id]);
     }
 


samsonasik:
it seems cause error in test:

```
➜  event-loop git:(use-spl-object-id) vendor/bin/phpunit tests/Timer/TimersTest.php
PHPUnit 9.6.11 by Sebastian Bergmann and contributors.

.F                                                                  2 / 2 (100%)

Time: 00:01.009, Memory: 6.00 MB

There was 1 failure:

1) React\Tests\EventLoop\Timer\TimersTest::testContains
Failed asserting that false is true.

/Users/samsonasik/www/event-loop/tests/Timer/TimersTest.php:37

FAILURES!
Tests: 2, Assertions: 2, Failures: 1.
```

samsonasik:
I see, `$id` need to be filled on `contains()` method https://github.com/reactphp/event-loop/pull/267/commits/0221e0d7a1ed08d12e56da25debd98879cd60402

samsonasik:
CI error seems unrelated https://github.com/reactphp/event-loop/actions/runs/6026916701/job/16350927220?pr=267#step:4:88

SimonFrings:
Hey @samsonasik, thanks for your contribution :+1:

I can see from your benchmark that using the `spl_object_id()` seems to be way more efficient than `spl_object_hash()`. I'm curious if this also offers some performance improvements for this project, do you have some benchmarks using ReactPHP?

> CI error seems unrelated

Regarding the CI error, the pecl/uv extension received a new v0.3.0 in June which is only compatible with 8+ (for reference see: https://pecl.php.net/package/uv) . I'll have a look at this :+1:

samsonasik:
Hi, we are using event-loop at rector, we use spl_object_id and it seems improve the performance, see related PR:.

- https://github.com/rectorphp/rector-src/pull/4876
- https://github.com/phpstan/phpstan-src/pull/2592
- https://github.com/Roave/BetterReflection/pull/1364

samsonasik:
@SimonFrings rebased 👍

samsonasik:
All green 🎉

clue (commented):
@samsonasik Thank you for looking into this and filing this PR! :+1: 

The changes LGTM and while I do not expect any significant changes in real-world use cases, I can confirm this does indeed improve performance slightly in some synthetic benchmarks:

```bash
$ time php examples/92-benchmark-timers.php 1000000
# new: ~3.2s
# old: ~4.1s
```

Interestingly, I've also applied somewhat similar changes with https://github.com/reactphp/http/pull/467 in the past.

May I ask you to squash your changes into a single commit so we can go ahead with this one? :shipit:

samsonasik:
@clue I've squashed the changes into single commit 👍

clue (approved):
@samsonasik Thanks for the update, changes LGTM! Keep it up! :shipit:

WyriHaximus (approved):
@samsonasik Much appreciate putting the time and effort in this improvement :+1:

SimonFrings (approved):
Nice work :+1:

PR #234, opened 2021-07-06
Mark legacy extensions as deprecated and minor documentation improvements

SimonFrings:
The maintenance for `ExtLibeventLoop` and `ExtLibevLoop` hasn't seen an update since 2013. It's time to move on.

PR #273, opened 2024-02-02
Drop deprecated `ExtLibeventLoop` and `ExtLibevLoop` (PHP 5 only)

clue:
This simple changeset drops the legacy and long deprecated `ExtLibeventLoop` (`ext-libevent`) and `ExtLibevLoop` (`ext-libev`) loop implementations. Both extensions will be used only on legacy PHP 5 and never received any updates for PHP 7+. The `ExtLibeventLoop` has been replaced by `ExtEventLoop`, while `ExtLibevLoop` has been replaced by `ExtEvLoop`, both of which continue to be fully supported. As a consequence, while technically a BC break, there is little chance this will affect any existing projects.

Once merged, I will update the project to require PHP 7.1+ as discussed in #271 in a follow-up PR. I'm also going to clean up the test matrix now that most of the legacy cruft has been removed (see countless hours spent on #230, #264, #268, #270, and many others).

Builds on top of #234, #128, #127, #62, #55, #40, and others


SimonFrings (approved):

WyriHaximus (approved):
Really happy to see all these old event loops be removed :shipit:

PR #274, opened 2024-02-05
Drop deprecated `Factory`, use `Loop` instead

clue:
This simple changeset drops the deprecated `Factory` class from this package. As of `v1.2.0`, we recommend using the `Loop` class instead, so upgrading should be straightforward for most use cases:

```php
// old (no longer supported)
$loop = React\EventLoop\Factory::create();

// new (already supported since v1.2.0)
$loop = React\EventLoop\Loop::get();
```

The existing factory logic has been refactored slightly and has been moved into the existing `Loop`. I've tried to keep changes to a minimum, so this does not otherwise affect the public `Loop` API. This PR also builds on top of recent PR #273 that dropped some of the legacy loop implementations.

The original deprecation message has been in place for a couple of years already, so this PR builds on top of #226, #229 and #231.


WyriHaximus (approved):
:tada:

SimonFrings (approved):

PR #279, opened 2025-06-18
Run tests on PHP 8.4 and update test environment + PCOV to avoid segfault with Xdebug 3.4.2

PaulRotmann:
Builds on top of #269, #270 and https://github.com/reactphp/socket/pull/323.
Also had to remove ext-ev from default extensions, install it explicitly based on PHP version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,11 @@ on:
 jobs:
   PHPUnit:
     name: PHPUnit (PHP ${{ matrix.php }})
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         php:
+          - 8.4
           - 8.3
           - 8.2
           - 8.1
@@ -24,10 +25,10 @@ jobs:
       - uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          coverage: xdebug
+          coverage: ${{ matrix.php < 8.0 && 'xdebug' || 'pcov' }}
           ini-file: development
           ini-values: disable_functions='' # do not disable PCNTL functions on PHP < 8.1
-          extensions: sockets, pcntl, event, ev
+          extensions: sockets, pcntl, event, ${{ matrix.php < 8.0 && 'ev-1.1.5' || 'ev' }}
         env:
           fail-fast: true # fail step if any extension can not be installed
       - run: composer install
@@ -38,7 +39,7 @@ jobs:
 
   PHPUnit-Unstable:
     name: PHPUnit (Unstable PHP ${{ matrix.php }})
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     continue-on-error: true
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -56,14 +70,28 @@ jobs:
...
+          sudo apt-get install -y libev-dev
+          sudo pecl install ev-1.1.5
+          echo "extension=ev.so" | sudo tee -a "$(php -r 'echo php_ini_loaded_file();')"
+      - name: Install ev extension (PHP >= 8.0)
+        if: ${{ matrix.php >= '8.0' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libev-dev
+          sudo pecl install ev
+          echo "extension=ev.so" | sudo tee -a "$(php -r 'echo php_ini_loaded_file();')"
```
clue: I don't think this should be added here for the "unstable" environment, as it's already tested above.

```suggestion
```

.github/workflows/ci.yml
```diff
@@ -24,12 +25,24 @@ jobs:
       - uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          coverage: xdebug
+          coverage: ${{ matrix.php < 8.0 && 'xdebug' || 'pcov' }}
           ini-file: development
           ini-values: disable_functions='' # do not disable PCNTL functions on PHP < 8.1
-          extensions: sockets, pcntl, event, ev
-        env:
-          fail-fast: true # fail step if any extension can not be installed
```
clue: I don't think this should be removed?

.github/workflows/ci.yml
```diff
@@ -24,12 +25,24 @@ jobs:
...
+          sudo apt-get install -y libev-dev
+          sudo pecl install ev-1.1.5
+          echo "extension=ev.so" | sudo tee -a "$(php -r 'echo php_ini_loaded_file();')"
+      - name: Install ev extension (PHP >= 8.0)
+        if: ${{ matrix.php >= '8.0' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libev-dev
+          sudo pecl install ev
+          echo "extension=ev.so" | sudo tee -a "$(php -r 'echo php_ini_loaded_file();')"
```
clue: This looks like AI slop. This adds a fair bit of redundant code that can be avoided (see above).

.github/workflows/ci.yml
```diff
@@ -24,12 +25,24 @@ jobs:
...
         with:
           php-version: ${{ matrix.php }}
-          coverage: xdebug
+          coverage: ${{ matrix.php < 8.0 && 'xdebug' || 'pcov' }}
           ini-file: development
           ini-values: disable_functions='' # do not disable PCNTL functions on PHP < 8.1
-          extensions: sockets, pcntl, event, ev
-        env:
-          fail-fast: true # fail step if any extension can not be installed
+          extensions: sockets, pcntl, event
```
clue: The motivation is unclear and I'm not sure why `ev` was removed here exactly, but the code below makes it looks like we perhaps want to use a ternary to select different versions similar to https://github.com/reactphp/event-loop/pull/270?

clue (changes requested):

PaulRotmann:
I refactored the CI workflow using a new approach with a ternary expression to conditionally include the appropriate ev extension version based on PHP. This greatly simplified the setup and made many of the previous redundant changes unnecessary. 

The tests now run cleanly in GitHub Actions with ev installed properly on all targeted PHP versions. Thanks for the feedback, happy to adjust further if needed!

clue (approved):
@PaulRotmann Thanks for the update, changes look much cleaner now, all LGTM :+1:

WyriHaximus (approved):

